PR #24, opened 2019-07-09
On 2047 full permissions contact on the register during apply

arshedmust:


alandburns:
should not be connecting directly to db

PR #32, opened 2019-07-22
ON-2120 : Add logging for all xxxApiClient errors

martin-hornsby:


martin-hornsby:
I'll raise a tech debt ticket later. It's basically due to the fact there's two end points for downloading files. I'll check with Dave before going further on my crusade 🏇 ⚔️

PR #44, opened 2019-08-28
Apr 671 remove structuremap

David-B-Read:


src/SFA.DAS.AdminService.Web/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace SFA.DAS.AdminService.Web.Extensions
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static void RegisterAllTypes<T>(this IServiceCollection services, Assembly[] assemblies,
+            ServiceLifetime lifetime = ServiceLifetime.Transient)
+        {
+            var typesFromAssemblies = assemblies.SelectMany(a => a.DefinedTypes.Where(x => x.GetInterfaces().Contains(typeof(T))));
+            foreach (var type in typesFromAssemblies)
```
davidgouge: We had to change this line slightly to `foreach (var type in typesFromAssemblies.Where(t=>!t.IsAbstract))` because we had an abstract class that implemented IValidator without a default ctor.  YMMV. 👍

src/SFA.DAS.AdminService.Web/Startup.cs
```diff
@@ -106,51 +105,83 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+
+            services.AddTransient<IAzureApiClient>(x => new AzureApiClient(
+                ApplicationConfiguration.AzureApiAuthentication.ApiBaseAddress,
+                x.GetService<IAzureTokenService>(),
+                x.GetService<ILogger<AzureApiClientBase>>(),
+                x.GetService<IWebConfiguration>(),
+                x.GetService<IOrganisationsApiClient>(),
+                x.GetService<IContactsApiClient>()));
+
+            services.AddTransient<CacheService, CacheService>();
```
davidgouge: You should be able to do `services.AddTransient<CacheService>();` for these types of registrations.

src/SFA.DAS.AdminService.Web/Extensions/ServiceCollectionExtensions.cs
David-B-Read: We're not using this yet, but I've updated it anyway, for future use

src/SFA.DAS.AdminService.Web/Startup.cs
David-B-Read: Have updated to the simplified versions

David-B-Read:
> Approved, but's what's with all the deletions?

The Assessor external API is duplicated from the das-assessor-service repo. I chose to delete the files rather than go through and migrate from StructureMap :D

PR #58, opened 2019-10-02
ON-2313 : Grade FHA Rework

martin-hornsby:


src/SFA.DAS.AssessorService.ApplyTypes/FinancialGrade.cs
```diff
@@ -9,10 +9,6 @@ public class FinancialGrade
         public string SelectedGrade { get; set; }
         public string InadequateMoreInformation { get; set; }
 
-        public FinancialDueDate OutstandingFinancialDueDate { get; set; }
-        public FinancialDueDate GoodFinancialDueDate { get; set; }
-        public FinancialDueDate SatisfactoryFinancialDueDate { get; set; }
```
martin-hornsby: These are ViewModel classes and not part of the schema. Hence the major upheaval in the code changes above ☝️

PR #69, opened 2019-11-07
Handle Answer Value to Array of String from String

alandburns:


src/SFA.DAS.AdminService.Web/Views/Apply/Applications/Section.cshtml
```diff
@@ -220,7 +220,7 @@
                                         }
                                         else
                                         {
-                                            if (@answer.Value != ",,")
+                                            if (@answer.Value != null && @answer.Value.Any())
                                             {
                                                  <p class="govuk-body">@answer.Value</p>
```
davidgouge: This needs to be `@answer.Value[0]` for simple answers but would need to loop through the array for Checkboxlist answers.

alandburns: Thanks. Yes that's what I said!

PR #71, opened 2019-11-12
APR-999 Roatp Gateway Assessor (GAC) dashboard

chrisjensenuk:
- Added new Role 'GAC'
- Using route names for navigation so we're not using controller/action magic strings.
- Added `/Views/RoatpAssessor/{1}/{0}.cshtml` as a view location so we can use convention for view locations.

### DIFF
--- a/src/SFA.DAS.AdminService.Web/Controllers/AccountController.cs
+++ b/src/SFA.DAS.AdminService.Web/Controllers/AccountController.cs
@@ -40,9 +40,17 @@ namespace SFA.DAS.AdminService.Web.Controllers
                 return RedirectToAction("InvalidRole", "Home");
             }
 
-            if (HttpContext.User.HasRoatpRoleOnly())
+            if (HttpContext.User.HasRoatpRolesOnly())
             {
-                return RedirectToAction("Index", "RoatpHome");
+                if (User.IsInRole(Roles.RoatpAssessorGateway))
+                {
+                    return RedirectToRoute(RouteNames.RoatpDashboard_Index_Get);
+                }
+
+                if (User.IsInRole(Roles.RoatpGatewayTeam))
+                {
+                    return RedirectToAction("Index", "RoatpHome");
+                }
             }
 
             return RedirectToAction("Index", "Dashboard");
--- a/src/SFA.DAS.AdminService.Web/Controllers/AccountController.cs
+++ b/src/SFA.DAS.AdminService.Web/Controllers/AccountController.cs
@@ -29,9 +29,9 @@ namespace SFA.DAS.AdminService.Web.Controllers
         [HttpGet]
         public IActionResult PostSignIn()
         {
-            if(!HttpContext.User.HasValidRole())
+            if(!User.HasValidRole())
             {
-                _logger.LogInformation($"PostSignIn - User '{HttpContext.User.Identity.Name}' does not have a valid role");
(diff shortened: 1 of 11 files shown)


src/SFA.DAS.AdminService.Web/Controllers/AccountController.cs
```diff
@@ -40,9 +40,17 @@ public IActionResult PostSignIn()
                 return RedirectToAction("InvalidRole", "Home");
             }
 
-            if (HttpContext.User.HasRoatpRoleOnly())
+            if (HttpContext.User.HasRoatpRolesOnly())
```
martin-hornsby: Please chop the `HttpContext` bit so it's consistent with the `User` calls below

src/SFA.DAS.AdminService.Web/RouteNames.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.AdminService.Web
+{
+    public static class RouteNames
+    {
```
martin-hornsby: Could you explain why you have these? They normally belong in the Controller and the HtmlHelper, TagHelper, etc will pick them up

src/SFA.DAS.AdminService.Web/Startup.cs
```diff
@@ -83,6 +84,12 @@ public void ConfigureServices(IServiceCollection services)
                 options.SupportedCultures = new List<CultureInfo> { new CultureInfo("en-GB") };
                 options.RequestCultureProviders.Clear();
             });
+
+            services.Configure<RazorViewEngineOptions>(o =>
+            {
+                o.ViewLocationFormats.Add("/Views/RoatpAssessor/{1}/{0}" + RazorViewEngine.ViewExtension);
```
martin-hornsby: Could you place a comment to say why you're doing stuff to the View Location schemes... this could have unwanted side affects

src/SFA.DAS.AdminService.Web/Controllers/AccountController.cs
chrisjensenuk: 👍

src/SFA.DAS.AdminService.Web/RouteNames.cs
chrisjensenuk: By using a const it makes refactoring a lot easier should the routes change.  Its easier to see where dependencies are.  With magic strings to controller/action its too easy to break links should an endpoint be moved.

src/SFA.DAS.AdminService.Web/Startup.cs
chrisjensenuk: The thinking is it means in my views I can use convention to return the required view and keep the controllers terse. So I can go:
`Return View();`
instead of
`Return View("~/Views/RoatpAssessor/Gateway/Index.cshtml");`

The location is added to the end of this list so its the last place to to be checked. All other views will have been resolved by then.

Yep I'm specifying execution root.  I could have added tilde so its more portable.

I'm not wedded to this approach so I'll remove as it has caused confusion. 👍

src/SFA.DAS.AdminService.Web/RouteNames.cs
martin-hornsby: I'll let it go, normal convention tends to be magic strings or blank

PR #79, opened 2019-12-03
CON-1127 : Fixed Edit standard within EPAO listing

VasanthaKasirajan3008:


VasanthaKasirajan3008:
Removed method GetEpaOrganisationContacts and Unit Test Added for the work  done for this fix.

src/SFA.DAS.AdminService.Web.Tests/Controllers/Register/RegisterBase.cs
```diff
@@ -73,12 +74,22 @@ public RegisterBase()
                 new OrganisationStandardSummary { Id = 1}
             };
 
+            List<DeliveryArea> deliveryAreas = new List<DeliveryArea>
+            {
+                new DeliveryArea { Id = 1 }
+            };
+
+            var organisationStandard = new OrganisationStandard() { OrganisationId = OrganisationOneOrganisationId };
```
ChrisJWoodcock: Not strictly necessary but this would make more sense if the OrganisationStandard.Id was set to organisationStandardId

src/SFA.DAS.AdminService.Web.Tests/Controllers/Register/RegisterControllerTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+
+        [Test]
+        public async Task GivenEditOrganisationStandard_WhenEditingOrganisationStandard_ThenReturnContacts()
+        {
+            //Act
+            var actionResult = await Sut.EditOrganisationStandard(organisationStandardId);
+
+            //Assert
```
ChrisJWoodcock: Could also Verify that the ContactApi was called here

PR #80, opened 2019-12-06
CON-1154 Import learner details request, keep duplicated code in sync with its source in assessor

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Domain/Entities/Ilr.cs
```diff
@@ -28,5 +28,11 @@ public class Ilr
         public int? CompletionStatus { get; set; }
         public long? EventId { get; set; }
         public DateTime? PlannedEndDate { get; set; }
+        public string DelLocPostCode { get; set; }
```
cofaulco: What are DelLocPostcode and AchDate ?

PR #93, opened 2020-02-03
APR-1268 - Dummy RoATP Gateway Review

martin-hornsby:


src/SFA.DAS.AssessorService.ApplyTypes/Roatp/GatewayApplicationSummaryItem.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace SFA.DAS.AssessorService.ApplyTypes.Roatp
+{
+    public class GatewayApplicationSummaryItem
```
David-B-Read: As commented on the Apply side, can we re-use this across Gateway, FHA and Assessor or do we want to store different information for different reviewer types?

martin-hornsby: Make this generic so we now have RoatpApplicationSummaryItem

PR #95, opened 2020-02-10
APR-1278 - RoATP Apply Snapshot

martin-hornsby:


src/SFA.DAS.AdminService.Web/Infrastructure/IRoatpApplicationApiClient.cs
```diff
@@ -36,6 +36,14 @@ public interface IRoatpApplicationApiClient
         Task EvaluateGateway(Guid applicationId, bool isGatewayApproved, string evaluatedBy);
 
         Task StartAssessorReview(Guid applicationId, string reviewer);
+
+        Task<Guid> SnapshotApplication(Guid Id, Guid NewApplicationId, List<RoatpApplySequence> sequences);
+    }
+
+    public class SnapshotApplicationRequest
```
MarkFCain: should be in own file?

martin-hornsby: Moved into ApplyTypes

PR #104, opened 2020-03-11
APR-1302 Organisation status check

MakCom:


src/SFA.DAS.AdminService.Web/Views/Roatp/Apply/Gateway/Application.cshtml
```diff
@@ -6,6 +6,17 @@
 @{
     ViewBag.Title = "Gateway evaluation";
     Layout = "_Layout";
+
```
David-B-Read: Might be better to have this logic outside the view so it can be unit tested

src/SFA.DAS.AdminService.Web/ViewModels/Roatp/Gateway/OrganisationStatusViewModel.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace SFA.DAS.AdminService.Web.ViewModels.Roatp.Gateway
+{
+    public class OrganisationStatusViewModel : RoatpGatewayPageViewModel
+    {
+        public DateTime? ApplicationSubmittedOn { get; set; }
+        public DateTime SourcesCheckedOn { get; set; }
+
+        public string UkrlpData { get; set; }
```
David-B-Read: A 'status' suffix might be more descriptive for these properties

src/SFA.DAS.AdminService.Web/Views/Roatp/Apply/Gateway/Application.cshtml
MakCom: We need to remove the logic altogether.

src/SFA.DAS.AdminService.Web/ViewModels/Roatp/Gateway/OrganisationStatusViewModel.cs
MakCom: Renaming is done

PR #114, opened 2020-03-19
fix overview

MarkFCain:


src/SFA.DAS.AdminService.Web/Services/Gateway/GatewayOverviewOrchestrator.cs
```diff
@@ -199,11 +197,11 @@ public async Task<RoatpGatewayApplicationViewModel> Handle(GetApplicationOvervie
                 {
                     var OfficeForStudentPage = await _qnaApiClient.GetPageBySectionNo(request.ApplicationId, 1, 5, "235");
```
David-B-Read: Would like to see constants replacing magic numbers here, unless we're moving it into the proposed config structure

David-B-Read: In fact, based on this morning's discussions, these calls to QnA API should be moved to Apply API

PR #115, opened 2020-03-20
Gateway submaster fix organisation check

MarkFCain:


src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayOrganisationChecksController.cs
```diff
@@ -110,5 +110,44 @@ public async Task<IActionResult> EvaluateTradingNamePage(TradingNamePageViewMode
...
+
+            _logger.LogInformation($"RoatpGatewayController-EvaluateOrganisationStatusPage-SubmitGatewayPageAnswer - ApplicationId '{viewModel.ApplicationId}' - PageId '{viewModel.PageId}' - Status '{viewModel.Status}' - UserName '{username}' - Comments '{comments}'");
+            try
+            {
+                await _applyApiClient.SubmitGatewayPageAnswer(viewModel.ApplicationId, viewModel.PageId, viewModel.Status, username, comments);
+            }
+            catch (Exception ex)
+            {
+                // MFCMFC Shutter page? throw again?
+                _logger.LogError(ex, "RoatpGatewayController-EvaluateTradingNamePage - SubmitGatewayPageAnswer - Error: '" + ex.Message + "'");
```
David-B-Read: Throw, and let the generic error page handle it, would be my suggestion

PR #116, opened 2020-03-23
Fix lifting logic from admin to apply

MarkFCain:


src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayOrganisationChecksController.cs
```diff
@@ -23,6 +23,7 @@ public class RoatpGatewayOrganisationChecksController : RoatpGatewayControllerBa
         private readonly ILogger<RoatpGatewayOrganisationChecksController> _logger;
         private readonly IGatewayOrganisationChecksOrchestrator _orchestrator;
 
+        private string GatewayViewsLocation = "~/Views/Roatp/Apply/Gateway/pages/";
```
David-B-Read: Is this intended to be used in the controller actions below?

src/SFA.DAS.AdminService.Web.Tests/Services/Gateway/Overview/Orchestrator/OverviewTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+
+            var request = new GetApplicationOverviewRequest(applicationId, UserName);
+
+            var response = _orchestrator.GetOverviewViewModel(request);
+
+            var viewModel = response.Result;
+
+
+            // This is lifted from elsewhere, but you can see the principle------
```
martin-hornsby: Not asserting here but may be intentional judging by the comment

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayOrganisationChecksController.cs
```diff
@@ -23,6 +23,7 @@ public class RoatpGatewayOrganisationChecksController : RoatpGatewayControllerBa
         private readonly ILogger<RoatpGatewayOrganisationChecksController> _logger;
         private readonly IGatewayOrganisationChecksOrchestrator _orchestrator;
 
+        private const string GatewayViewsLocation = "~/Views/Roatp/Apply/Gateway/pages/";
```
martin-hornsby: Could it be done without the trailing slash. Looks a bit un-natural when viewing the paths to the view down below

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayOrganisationChecksController.cs
```diff
@@ -53,7 +54,7 @@ public async Task<IActionResult> EvaluateLegalNamePage(LegalNamePageViewModel vi
             if (validationResponse.Errors != null && validationResponse.Errors.Any())
             {
                 viewModel.ErrorMessages = validationResponse.Errors;
-                return View("~/Views/Roatp/Apply/Gateway/pages/LegalName.cshtml", viewModel);
+                return View($"{GatewayViewsLocation}LegalName.cshtml", viewModel);
```
martin-hornsby: As mention above, splitting out the training slash makes it more readable. So it would become:
`return View($"{GatewayViewsLocation}/LegalName.cshtml", viewModel);`

src/SFA.DAS.AdminService.Web/Services/Gateway/GatewayOverviewOrchestrator.cs
```diff
@@ -68,13 +68,13 @@ public async Task<RoatpGatewayApplicationViewModel> GetOverviewViewModel(GetAppl
...
-                        new GatewaySection { SectionNumber = 2, PageId = "1-20", UrlTag = "trading-name",LinkTitle = "Trading name", HiddenText = "", Status = "" },
-                        new GatewaySection { SectionNumber = 3, PageId = "1-30", UrlTag = "organisation-status", LinkTitle = "Organisation status", HiddenText = "", Status = "" },
-                        new GatewaySection { SectionNumber = 4, PageId = "1-40", UrlTag="Address",LinkTitle = "Address", HiddenText = "", Status = "" },
-                        new GatewaySection { SectionNumber = 5, PageId = "1-50", UrlTag = "ico-number",LinkTitle = "ICO registration number", HiddenText = "", Status = "" },
-                        new GatewaySection { SectionNumber = 6, PageId = "1-60", UrlTag="website",  LinkTitle = "Website address", HiddenText = "", Status = "" },
-                        new GatewaySection { SectionNumber = 7, PageId = "1-70", UrlTag="organisation-risk", LinkTitle = "Organisation high risk", HiddenText = "", Status = "" }
+                        new GatewaySection { SectionNumber = 1, PageId = "LegalName",  LinkTitle = "Legal name", HiddenText = "", Status = "" },
```
martin-hornsby: Could you use GatewayPageIds.LegalName here rather than magic string?

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayOrganisationChecksController.cs
MarkFCain: agree

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpApplicationApiClient.cs
```diff
@@ -221,7 +248,18 @@ private async Task<T> Get<T>(string uri)
             }
         }
 
-        
+        private async Task<string> Get(string uri)
```
martin-hornsby: Is it a good idea to return string from a REST API end point? Usually it would be JSON

PR #117, opened 2020-03-23
APR-1303 - Address check

MakCom:


src/SFA.DAS.AssessorService.ApplyTypes/Roatp/RoatpQnaConstants.cs
```diff
@@ -20,6 +20,12 @@ public static class Preamble
 
                 public static class QuestionIds
                 {
+                    public static string UKRLPLegalAddressLine1 = "PRE-40";
```
David-B-Read: Are these needed? Gateway calls to QnA should be contained to Apply API

MakCom: No. They are not needed anymore. I have removed them. I initially left them in case they might be needed or the whole class being removed (which I think will be the end result of getting all QnA data via ApplyService).

PR #123, opened 2020-03-26
APR-1322 Website address

MakCom:


src/SFA.DAS.AdminService.Web/Services/Gateway/GatewayOrganisationChecksOrchestrator.cs
```diff
@@ -227,5 +227,53 @@ public async Task<IcoNumberViewModel> GetIcoNumberViewModel(GetIcoNumberRequest
...
+            }
+            else
+            {
+                model.SubmittedWebsite = ukrlpWebsite;
+            }
+           
+            var ukrlpDetails = await _applyApiClient.GetUkrlpDetails(request.ApplicationId);
+            if (ukrlpDetails != null && ukrlpDetails.ContactDetails != null)
+            {
+                model.UkrlpWebsite = ukrlpDetails.ContactDetails.FirstOrDefault(x => x.ContactType == RoatpGatewayConstants.ProviderContactDetailsTypeLegalIdentifier).ContactWebsiteAddress;
```
David-B-Read: What if there isn't a legal address returned by UKRLP? Would this be a null reference?

MakCom: Check for null reference  (?) added

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayOrganisationChecksController.cs
```diff
@@ -220,5 +220,41 @@ public async Task<IActionResult> EvaluateIcoNumberPage(IcoNumberViewModel viewMo
 
             return RedirectToAction("ViewApplication", "RoatpGateway", new { viewModel.ApplicationId });
         }
+
+        [HttpGet("/Roatp/Gateway/{applicationId}/Page/WebsiteAddress")]
+        public async Task<IActionResult> GetWebsitePage(Guid applicationId)
+        {
+            var username = _contextAccessor.HttpContext.User.UserDisplayName();
+            var viewModel = await _orchestrator.GetWebsiteViewModel(new GetWebsiteRequest(applicationId, username));
+            return View("~/Views/Roatp/Apply/Gateway/pages/Website.cshtml", viewModel);
```
MarkFCain: There's a variable GatewayViewsLocation in the gateway-submaster that you can use instead of the long path

MakCom: Long path replaced with variable

src/SFA.DAS.AdminService.Web/Services/Gateway/GatewayOrganisationChecksOrchestrator.cs
```diff
@@ -227,5 +227,53 @@ public async Task<IcoNumberViewModel> GetIcoNumberViewModel(GetIcoNumberRequest
...
+                var applyWebsite = await _applyApiClient.GetWebsiteAddressManuallyEntered(request.ApplicationId);
+                if (!string.IsNullOrEmpty(applyWebsite))
+                {
+                    model.SubmittedWebsite = applyWebsite;
+                }
+            }
+            else
+            {
+                model.SubmittedWebsite = ukrlpWebsite;
+            }
```
MichaelYoung1981: I'd have thought it would make sense for this logic to go in Apply so me can keep admin as dumb as possible.

David-B-Read: If moving this API-side, I'd recommend returning the source (UKRLP/Apply) in the data returned

MakCom: I'll move the logic to ApplyAPI, but why we need the source  (UKRLP/Apply) ? We don't make use of it. Please, confirm that it needs to be done @David-B-Read

David-B-Read: Logging API side to indicate where the website was retrieved from would be fine instead

MakCom: Sure, will do that. Thanks!

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/OrganisationChecks/WebsiteAddressTests.cs
```diff
@@ -0,0 +1,139 @@
...
+        public void check_website_request_is_sent()
+        {
+            var applicationId = Guid.NewGuid();
+
+            _orchestrator.Setup(x => x.GetWebsiteViewModel(new GetWebsiteRequest(applicationId, username)))
+                .ReturnsAsync(new WebsiteViewModel())
+                .Verifiable("view model not returned");
+
+            var _result = _controller.GetWebsitePage(applicationId).Result;
+            _orchestrator.Verify(x => x.GetWebsiteViewModel(It.IsAny<GetWebsiteRequest>()), Times.Once());
```
MichaelYoung1981: All .IsAnys need to be replaced with expected values.

MakCom: All .IsAnys are replaced with the expected values now

PR #142, opened 2020-04-03
Apr 1311 external apis down shutter page

David-B-Read:


martin-hornsby:
Presuming `ExternalApiException.cs` will get replaced with my Exception once APR-1551 is pulled in

David-B-Read:
> Presuming `ExternalApiException.cs` will get replaced with my Exception once APR-1551 is pulled in

That exception type is to denote API calls that have a dependency on services external to ESFA, and will trigger the shutter page. Exceptions from our own services will be handled as normal, i.e. regular service unavailable page

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpClients/RoatpApplicationApiClient.cs
```diff
@@ -153,7 +153,15 @@ public async Task<GatewayPageAnswer> GetGatewayPageAnswer(Guid applicationId, st
 
         public async Task<GatewayCommonDetails> GetPageCommonDetails(Guid applicationId, string pageId, string userName)
         {
-            return await Get<GatewayCommonDetails>($"Gateway/Page/CommonDetails/{applicationId}/{pageId}/{userName}");
+            try
+            {
+                return await Get<GatewayCommonDetails>($"Gateway/Page/CommonDetails/{applicationId}/{pageId}/{userName}");
+            }
+            catch (Exception ex)
```
MichaelYoung1981, line 160: Should all exceptions result in the same shutter page?  There are parts of this call which aren't related to external apis,

MichaelYoung1981: Might it be better for the apply api to return a particular response code (service unavailable maybe?) when an external API is down and show the error page based on that instead?

David-B-Read: I went for exceptions rather than response codes, because we'd need to handle exceptions anyway for timeouts, i.e. the external service 4 hops away doesn't respond in time

MichaelYoung1981: Fair enough.  Can you merge in @martin-hornsby 's branch and catch the new exception thrown?  That way and non-api exceptions (not that there are ever likely to be any!) will be handled in the default way.

David-B-Read: Done

martin-hornsby:
1551 has been merged into master

PR #143, opened 2020-04-06
Apr 1343 people in control high risk

MarkFCain:


src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        [SetUp]
+        public void Setup()
+        {
+            _applyApiClient = new Mock<IRoatpApplicationApiClient>();
+            _contextAccessor = new Mock<IHttpContextAccessor>();
+            _gatewayValidator = new Mock<IRoatpGatewayPageViewModelValidator>();
+            _logger = new Mock<ILogger<RoatpGatewayPeopleInControlController>>();
+            _orchestrator = new Mock<IPeopleInControlOrchestrator>();
+
+            var user = new ClaimsPrincipal(new ClaimsIdentity(new Claim[]
```
David-B-Read, line 42: Just a heads up, I think this setup has been moved into a test base class now - might not be merged in yet

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            var vm = new PeopleInControlHighRiskPageViewModel
+            {
+                Status = SectionReviewStatus.Pass,
+                SourcesCheckedOn = DateTime.Now,
+                ErrorMessages = new List<ValidationErrorDetail>()
+            };
+
+            vm.SourcesCheckedOn = DateTime.Now;
+
+            _applyApiClient.Setup(x =>
```
MichaelYoung1981, line 90: This setup isn't needed, the verify should take care of it.

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                x.SubmitGatewayPageAnswer(applicationId, pageId, vm.Status, username, It.IsAny<string>()));
+
+            var result = _controller.EvaluatePeopleInControlHighRiskPage(vm).Result;
+
+            _applyApiClient.Verify(x => x.SubmitGatewayPageAnswer(It.IsAny<Guid>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>()), Times.Once);
+            _orchestrator.Verify(x => x.GetPeopleInControlHighRiskViewModel(It.IsAny<GetPeopleInControlHighRiskRequest>()), Times.Once());
```
MichaelYoung1981, line 96: Replace IsAny with specific values.

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                .Verifiable("view model not returned");
+
+
+            _applyApiClient.Setup(x =>
+                x.SubmitGatewayPageAnswer(applicationId, pageId, viewModel.Status, username, It.IsAny<string>()));
+
+            var result = _controller.EvaluatePeopleInControlHighRiskPage(viewModel).Result;
+
+            _applyApiClient.Verify(x => x.SubmitGatewayPageAnswer(It.IsAny<Guid>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>()), Times.Never);
+            _orchestrator.Verify(x => x.GetPeopleInControlHighRiskViewModel(It.IsAny<GetPeopleInControlHighRiskRequest>()), Times.Once());
```
MichaelYoung1981, line 138: Replace IsAny with specific check.

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            viewModel.ApplicationId = applicationId;
+            viewModel.PageId = viewModel.PageId;
+            viewModel.SourcesCheckedOn = DateTime.Now;
+
+            _orchestrator.Setup(x => x.GetPeopleInControlHighRiskViewModel(It.IsAny<GetPeopleInControlHighRiskRequest>()))
+                .ReturnsAsync(viewModel)
+                .Verifiable("view model not returned");
+
+
+            _applyApiClient.Setup(x =>
```
MichaelYoung1981, line 132: Setup isn't needed as it isn't returning anything,

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                        new ValidationErrorDetail {Field = "OptionFail", ErrorMessage = "needs text"}
+                        }
+                }
+                );
+
+            viewModel.ApplicationId = applicationId;
+            viewModel.PageId = viewModel.PageId;
+            viewModel.SourcesCheckedOn = DateTime.Now;
+
+            _orchestrator.Setup(x => x.GetPeopleInControlHighRiskViewModel(It.IsAny<GetPeopleInControlHighRiskRequest>()))
```
MichaelYoung1981, line 127: Replace IsAny with specific check

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
```diff
@@ -48,6 +48,28 @@ public async Task<IActionResult> EvaluatePeopleInControlPage(PeopleInControlPage
...
+
+        [HttpPost("/Roatp/Gateway/{applicationId}/Page/PeopleInControlRisk")]
+        public async Task<IActionResult> EvaluatePeopleInControlHighRiskPage(PeopleInControlHighRiskPageViewModel viewModel)
+        {
+            var username = _contextAccessor.HttpContext.User.UserDisplayName();
+            var vmRebuild = await _orchestrator.GetPeopleInControlHighRiskViewModel(new GetPeopleInControlHighRiskRequest(viewModel.ApplicationId, username));
+            viewModel.CompanyDirectorsData = vmRebuild?.CompanyDirectorsData;
+            viewModel.PscData = vmRebuild?.PscData;
+            viewModel.TrusteeData = vmRebuild?.TrusteeData;
+            viewModel.WhosInControlData = vmRebuild?.WhosInControlData;
```
MichaelYoung1981, line 69: Is this only required when there are validation errors?  I wonder if it's be better creating an overload of SubmitGatewayPageAnswer so the call is only made if there are errors rather than all the time.

src/SFA.DAS.AdminService.Web.Tests/Controllers/Gateway/PeopleInControl/PeopleInControlHighRiskTests.cs
MarkFCain: Thanks for heads up, am changing test to use this

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
MarkFCain: This is a good idea,  Want to discuss in DevCuddle maybe, and figure out how to implement (this story? tech task?)

MarkFCain: I'll push this branch as is, and investigate this in its own branch or something like that

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayControllerBase.cs
```diff
@@ -57,13 +59,24 @@ public string SetupGatewayPageOptionTexts(RoatpGatewayPageViewModel viewModel)
...
+            return await SubmitGatewayPageAnswer(viewModel, errorViewName, validationResponse.Errors);
+        }
+
+        protected async Task<IActionResult> SubmitGatewayPageAnswer(RoatpGatewayPageViewModel viewModel, string errorViewName,  List<ValidationErrorDetail> validationErrors)
+        {
+            if (validationErrors == null)
+            {    
+                var validationResponse = await GatewayValidator.Validate(viewModel);
+                validationErrors = validationResponse.Errors;
+            }
```
MichaelYoung1981, line 75: I'd remove this block...if they're calling this overload it'd be safe to assume validation has been done.

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
```diff
@@ -39,15 +39,50 @@ public async Task<IActionResult> GetGatewayPeopleInControlPage(Guid applicationI
             [HttpPost("/Roatp/Gateway/{applicationId}/Page/PeopleInControl")]
             public async Task<IActionResult> EvaluatePeopleInControlPage(PeopleInControlPageViewModel viewModel)
             {
+                var validationCheck = await GatewayValidator.Validate(viewModel);
+
+                if (validationCheck.Errors == null || !validationCheck.Errors.Any())
+                    return await SubmitGatewayPageAnswer(viewModel, $"{GatewayViewsLocation}/PeopleInControl.cshtml",
```
MichaelYoung1981, line 45: I don't think this should call Submit when it knows it's just going to fail.

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
```diff
@@ -39,15 +39,50 @@ public async Task<IActionResult> GetGatewayPeopleInControlPage(Guid applicationI
...
+
                 var username = _contextAccessor.HttpContext.User.UserDisplayName();
-                var vmRebuild = await _orchestrator.GetPeopleInControlViewModel(new GetPeopleInControlRequest(viewModel.ApplicationId, username));
+                var vmRebuild =
+                    await _orchestrator.GetPeopleInControlViewModel(
+                        new GetPeopleInControlRequest(viewModel.ApplicationId, username));
                 viewModel.CompanyDirectorsData = vmRebuild?.CompanyDirectorsData;
                 viewModel.PscData = vmRebuild?.PscData;
                 viewModel.TrusteeData = vmRebuild?.TrusteeData;
                 viewModel.WhosInControlData = vmRebuild?.WhosInControlData;
```
MichaelYoung1981, line 55: I'm a bit confused why this bit is here...won't the reloaded viewmodel properties just be discarded?  Submit will pass as the validation has already been done so why it the vmRebuild needed?

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
MarkFCain: yeah I addressed this anyway

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
MarkFCain: It isn't, it's calling it when there aren't any errors

MichaelYoung1981: Thanks, turns out I can't read. 😁

MarkFCain: Oh I think I got what you mean, you mean after it rebuilds the model. Ok.

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
MarkFCain: No the rebuild is being done because it failed validiation, and for this page these parts are not primitive types so can't be embedded in a hidden, so needs rebuilding.

MichaelYoung1981: Yeah, it's because I misread the above bit.  What'd probably make more sense in my head would be to do the rebuild of the VM if validation failed and then call SubmitGatewayPageAnswer once at the end regardless of whether validation passed or failed and let it do the redirection.

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayControllerBase.cs
MarkFCain: done

src/SFA.DAS.AdminService.Web/Controllers/Roatp/Apply/RoatpGatewayPeopleInControlController.cs
MarkFCain: It's a bit clearer now cos it's dropping into a View directly rather than indirectly (via the submit answer business)

PR #162, opened 2020-05-20
Apr 1634 shared code nuget

David-B-Read:


src/SFA.DAS.AssessorService.Api.Types/SFA.DAS.AssessorService.Api.Types.csproj
```diff
@@ -2,14 +2,15 @@
...
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="MediatR" Version="7.0.0" />
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\SFA.DAS.AdminService.Common\SFA.DAS.AdminService.Common.csproj" />
```
ChrisJWoodcock, line 13: The AssessorService.Api.Types is / was a duplication of the code in [das-assessor-service] it seems suspicious that this would have a reference to a library in [das-admin-service]

ChrisJWoodcock: Also the title of this PR is shared code nuget - what is shared and with who is it shared and would something which is shared be in admin service repo?

David-B-Read: The Nuget package is designed to be consumed by the das-roatp-assessor, das-roatp-gateway and das-roatp-oversight repos

David-B-Read: The reference is there because ValidationResponse has been moved into AdminService.Common

PR #167, opened 2020-06-10
CON-1834 Updated status for duplicate and amend of certificates with new print notification status.

ChrisJWoodcock:


src/SFA.DAS.AssessorService.Domain/Consts/CertificateStatus.cs
```diff
@@ -9,11 +11,39 @@ public static class CertificateStatus
...
+        }
+
+        public static bool HasPrintNotificateStatus(string certificateStatus)
+        {
+            return PrintNotificationStatus.Contains(certificateStatus);
+        }
+
+        public static bool CanRequestDuplicateCertificate(string status)
+        {
+            var allowed = PrintProcessStatus;
```
cofaulco, line 39: No need for the local variable "allowed"

PR #171, opened 2020-07-22
Con 2074 display print process batch status

hkf-tech:


src/SFA.DAS.AdminService.Web.Tests/Controllers/BatchSearch/BatchSearchControllerTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        public void Setup()
+        {
+            _staffBatchLogResult = SetupStaffBatchLogResult();
+            _staffBatchSearchResponse = SetUpBatchSearchResponse();
+            ApiClient apiClient = SetUpApiClient();
+            _controller = new BatchSearchController(Mock.Of<ILogger<BatchSearchController>>(), apiClient, Mock.Of<ISessionService>());
+        }
+
+        [Test]
+        public async Task batch_log_view_model_is_as_expected()
```
cofaulco, line 37: could test names be a little more specific? What is "as expected"?

PR #174, opened 2020-08-05
Con 1568

hkf-tech:


martin-hornsby:
This is just an observation, however I'd be very tempted to rename **ClientApiAuthentication**  to something more meaningful so we know that it points to the assessor service internal API.

If I recall the reason why **ApiAuthentication** used to exist was to distinguish the split between EPAO Applications and Assessor Service API. I'm guessing that split no longer needs to be there but it did keep useful separation of concerns.

src/SFA.DAS.AdminService.Settings/WebConfiguration.cs
```diff
@@ -5,10 +5,6 @@ namespace SFA.DAS.AdminService.Settings
 {
     public class WebConfiguration : IWebConfiguration
     {
-        public AuthSettings Authentication { get; set; }
-
-        [JsonRequired] public ApiAuthentication ApiAuthentication { get; set; }
-
         [JsonRequired] public AzureApiAuthentication AzureApiAuthentication { get; set; }
 
         [JsonRequired] public ClientApiAuthentication ClientApiAuthentication { get; set; }
```
ChrisJWoodcock, line 10: There are several used of ClientApiAuthentication - this one is for the Admin Web to connect to the Assessor Api - however it should be called EpaoApiAuthentication - as it is the equivalent of the RoatpApiAuthentication for Epao - and also because for some reason Roatp uses the term Assessor to actually mean the Admin service...

PR #179, opened 2020-08-19
CON-2233-add additional print status information.

shomavg:


ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Models/Search/LearnerDetailForStaffViewModel.cs
```diff
@@ -21,45 +15,46 @@ public class LearnerDetailForStaffViewModel
...
         {
-            if (learnerCertificateLogs.Any())
-            {
-                var learnerLog = learnerCertificateLogs.OrderByDescending(e => e.EventTime)
-                    .FirstOrDefault(l => l.Status == Learner.CertificateStatus);
-                return learnerLog?.EventTime.UtcToTimeZoneTime();
-            }
-
-            return null;
+            if (learnerCertificateStatus == "SentToPrinter")
```
ChrisJWoodcock, line 55: This should be any of the print notification status i.e. CertificateStatus.HasPrintNotificateStatus() method

src/SFA.DAS.AdminService.Web/Views/Search/Select.cshtml
```diff
@@ -183,7 +184,7 @@ else
                             @CertificateStatus.GetDisplayName(Model.Learner.CertificateStatus)
                         </dd>
                     </div>
-                    @if (!string.IsNullOrEmpty(Model.DateStatusTitle))
+                    @if (Model.UpdatedStatusDate != null)
```
ChrisJWoodcock, line 187: If will always have a date according to the logic above - so what is the if statement for - which AC?

ChrisJWoodcock (changes requested):

ChrisJWoodcock (dismissed):

ChrisJWoodcock (approved):

ChrisJWoodcock (approved):

PR #186, opened 2020-10-21
Con 2263: Updates for ScheduleRun LastRunStatus 

hkf-tech:


src/SFA.DAS.AdminService.Web/Views/ScheduleConfig/Restart.cshtml
```diff
@@ -0,0 +1,50 @@
...
+
+                <form method="post">
+                    @Html.HiddenFor(m => m.Id)
+                    @Html.HiddenFor(m => m.ScheduleType)
+                    <button class="govuk-button">Confirm</button>
+                </form>
+            }
+            else
+            {
+                <p class="govuk-body-l">No schedule found.</p>
```
cofaulco, line 45: why would there be "No schedule found" on a confirmation screen?

ChrisJWoodcock (approved):

src/SFA.DAS.AdminService.Web/Views/ScheduleConfig/Restart.cshtml
hkf-tech: I copied the pattern from delete confirmation :) .
The only way I can think that some one else deletes the schedule after the get operation

hkf-tech: done. Removed the if else block

cofaulco (approved):

PR #187, opened 2020-10-21
CON-2298 Epao Withdrawal

ChrisJWoodcock:


ChrisJWoodcock (commented):
All looks good

src/SFA.DAS.AdminService.Web/Views/Application/Returned.cshtml
```diff
@@ -1,9 +1,15 @@
-@using SFA.DAS.AdminService.Web.Helpers;
-
+@using SFA.DAS.AssessorService.Domain.Consts;
+@using SFA.DAS.AdminService.Web.Domain.Apply;
 @model SFA.DAS.AdminService.Web.ViewModels.Apply.Applications.ApplicationReturnedViewModel
 
 @{
-    ViewBag.Title = Model.WarningMessages.Any() ? "Issue during approval" : "Feedback sent";
+    ViewBag.Title = Model.SequenceNo == ApplyConst.ORGANISATION_WITHDRAWAL_SEQUENCE_NO || Model.SequenceNo == ApplyConst.STANDARD_WITHDRAWAL_SEQUENCE_NO
```
cofaulco, line 6: Would be better if this logic was done in the ViewModel itself?

cofaulco (approved):

src/SFA.DAS.AdminService.Web/Views/Application/Returned.cshtml
ChrisJWoodcock: We will need to have a TechDebt clean up to move logic to view models and add tests as there are hundred of examples of code like this

PR #206, opened 2020-12-10
Con 2755 application overview page

Howesy77:


src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
```diff
@@ -71,6 +95,17 @@
...
                         </div>
                     }
+                    @if (Model.IsWithdrawal)
+                    {
+                        <div class="govuk-summary-list__row">
+                            <dt class="govuk-summary-list__key">Contact details</dt>
+                            <dd class="govuk-summary-list__value">
+                                @Model.ContactName
+                                <br />
+                                <a href="mailto:@Model.ContactEmail" class="govuk-link">@Model.ContactEmail</a>
```
ChrisJWoodcock, line 105: There may be a thing about no using mailto - I've had this in the past query with deen or design team?

src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
```diff
@@ -36,27 +36,51 @@
...
                         </div>
                     }
-                    @if (!string.IsNullOrWhiteSpace(Model.LegalName))
+                    @if (!string.IsNullOrWhiteSpace(Model.Standard) && Model.SequenceNo == ApplyConst.STANDARD_WITHDRAWAL_SEQUENCE_NO)
                     {
                         <div class="govuk-summary-list__row">
-                            <dt class="govuk-summary-list__key">Registered name</dt>
-                            <dd class="govuk-summary-list__value">@Model.LegalName</dd>
+                            <dt class="govuk-summary-list__key">Details</dt>
+                            <dd class="govuk-summary-list__value">@Model.Standard</dd>
```
ChrisJWoodcock, line 57: It doesn't look like this includes the STRef - I think it should from the AC

src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
```diff
@@ -36,27 +36,51 @@
                             <dd class="govuk-summary-list__value">@Model.ApplicationReference</dd>
                         </div>
                     }
+                    @if (!string.IsNullOrWhiteSpace(Model.LegalName) && Model.IsWithdrawal)
+                    {
+                        <div class="govuk-summary-list__row">
+                            <dt class="govuk-summary-list__key">Organisation name</dt>
+                            <dd class="govuk-summary-list__value">@Model.LegalName</dd>
```
ChrisJWoodcock, line 43: This needs checking for correct name against the use on the tab view organisation name

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
Howesy77: Not sure - this is how it is in the design

Howesy77: Removed mailto:

src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
Howesy77: Changed to ST Ref

src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
Howesy77: Changed to EPOA Assessor Name

ChrisJWoodcock (changes requested):
There should also be similar change to the Sequence_ReadOnly.cshtml - maybe there are missing AC's - or Deen just assumed you would know

Howesy77:
> There should also be similar change to the Sequence_ReadOnly.cshtml - maybe there are missing AC's - or Deen just assumed you would know

Done

ChrisJWoodcock (approved):

PR #211, opened 2020-12-15
Update Assessment.cshtml

Howesy77:


src/SFA.DAS.AdminService.Web/Views/Application/Assessment.cshtml
```diff
@@ -19,7 +19,7 @@
 <main class="govuk-main-wrapper " id="main-content" role="main">
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-            <h1 class="govuk-heading-xl">Assessment summary</h1>
+            <h1 class="govuk-heading-xl">Complete review</h1>
```
ChrisJWoodcock, line 22: How come it is ok to change the heading on the page for both apply and withdrawal but the button and text and subtly different for each journey? 

Did you query that with Deen / Design?

Howesy77: Ok changed

ChrisJWoodcock (approved):

PR #218, opened 2021-01-07
Apr 2079 governance report download

chrisfoster76:


src/SFA.DAS.AdminService.Web.Tests/Controllers/Roatp/DownloadRoatpControllerTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        [Test]
+        public async Task ApplicationDownloadCsv_downloads_file()
+        {
+            var apiResponse = new List<RoatpApplicationOversightDownloadItem>();
+
+            _applicationApplyApiClient.Setup(
+                x => x.GetApplicationOversightDetailsForDownload(It.IsAny<DateTime>(), It.IsAny<DateTime>()))
+                .ReturnsAsync(apiResponse);
+
+            var expectedFileContents = Encoding.ASCII.GetBytes("THIS IS A TEST");
```
martin-lush, line 49: This is not valid CSV content

src/SFA.DAS.AdminService.Web/Models/Roatp/BooleanConverter.cs
```diff
@@ -0,0 +1,21 @@
...
+        public override string ConvertToString(object value, IWriterRow row, MemberMapData memberMapData)
+        {
+            if (value == null)
+            {
+                return string.Empty;
+            }
+
+            var boolValue = (bool)value;
+
+            return boolValue ? "Y" : "N";
```
martin-lush, line 18: how about bool.TryConvert as better solution? If value is empty then it'll return Y which may not be intended

src/SFA.DAS.AdminService.Web.Tests/Controllers/Roatp/DownloadRoatpControllerTests.cs
martin-lush, line 49: This is not valid CSV content, thus not a realistic test scenario

martin-lush (changes requested):

src/SFA.DAS.AdminService.Web.Tests/Controllers/Roatp/DownloadRoatpControllerTests.cs
chrisfoster76: The test is to verify that the output from the csv export service is what is returned from the action, not that the content is in a CSV format (which is a CSVHelper concern), so the test is valid.

src/SFA.DAS.AdminService.Web/Models/Roatp/BooleanConverter.cs
chrisfoster76: A null value returns empty string, but realistically this converter is for use with a bool so not really an issue?

src/SFA.DAS.AdminService.Web.Tests/Controllers/Roatp/DownloadRoatpControllerTests.cs
martin-lush: True, although it's better if the file contents reflected reality. Otherwise why go to the efforts of initializing & validating the contents below.

A better test would be to check that the result isn't null

martin-lush (dismissed):

MarkFCain (approved):

PR #237, opened 2021-04-30
[CON-3457] Add redis cache for dataprotection keys

cofaulco:


src/SFA.DAS.AdminService.Web/Startup.cs
```diff
@@ -109,24 +113,18 @@ public void ConfigureServices(IServiceCollection services)
                     o.ViewLocationFormats.Add("/Views/Application/{0}" + RazorViewEngine.ViewExtension);
                 });
 
-            services.AddSession(opt => 
+            services.AddSession(opt =>
             {
                 opt.IdleTimeout = TimeSpan.FromHours(1);
             });
 
-            if (!_env.IsDevelopment())
```
kola-samuel, line 117: why is this logic removed

kola-samuel (changes requested):
Response required

src/SFA.DAS.AdminService.Web/Startup.cs
cofaulco: _env is passed down and the check is done inside the extension to determine whether or not to use redis for both session caching and data protection keys.

kola-samuel (approved):

kola-samuel (approved):

VasanthaKasirajan3008 (approved):

ChrisJWoodcock (approved):

PR #241, opened 2021-05-28
SV-525-Ability to edit version

techtoniq:


src/SFA.DAS.AdminService.Web/Views/CertificateAmend/Version.cshtml
```diff
@@ -0,0 +1,105 @@
...
+else
+{
+    <a class="govuk-back-link" asp-controller="CertificateAmend" asp-action="Check" asp-route-certificateId=@Model.Id>Back</a>
+}
+
+<main class="govuk-main-wrapper " id="main-content" role="main">
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+
+            <form type="post" asp-method="Option" class="apprentice-option-form" action="" method="post" novalidate="novalidate">
```
CodeWolf86, line 21: Should the method and classes reference version instead?

src/SFA.DAS.AdminService.Web/Views/CertificateAmend/Version.cshtml
```diff
@@ -0,0 +1,105 @@
...
+                        </legend>
+
+                        <span id="options-conditional-hint" class="govuk-hint">
+                            The version will appear on the certificate.
+                        </span>
+
+
+
+
+
```
CodeWolf86, line 43: Unsure if all the whitespace here is git's doing or the file itself? can it be shrunk a bit?

CodeWolf86 (changes requested):
Few things on this PR to be tweaked i think
- We need to change the Options page to be as per the prototype with radio buttons
- The acceptance criteria whereby when you change version, if option is set, should be wiped out isn't functioning at the moment
- If you are modifying a standard with only 1 option ( Standard Code 417 ) i think the change option link is still visible on the check page.
- If you change version within a standard that only has 1 option, that option should be pre-selected.

src/SFA.DAS.AdminService.Web/Controllers/CertificateAmendController.cs
```diff
@@ -32,8 +31,12 @@ public async Task<IActionResult> Check(Guid certificateId, string searchString,
             certificateCheckViewModel.SearchString = searchString;
             certificateCheckViewModel.FromApproval = fromApproval;
 
+            var standards = await ApiClient.GetStandardVersions(certificateCheckViewModel.StandardCode);
+            certificateCheckViewModel.StandardHasMultipleVersions = standards.Count > 1;
+
             var options = await ApiClient.GetOptions(certificateCheckViewModel.StandardCode);
-            TempData["HideOption"] = !options.Any();
+            TempData["HideOption"] = false;
```
CodeWolf86, line 39: Can this HideOption temp data now be removed?

CodeWolf86 (changes requested):
Couple of comments on the PR, plus this PR to update based on the api removal: https://github.com/SkillsFundingAgency/das-admin-service/pull/244

Items I think are missing: 
When Version is not set on the certificate, it's not showing "None" it's showing a blank space. ( Can see this by standardUID being blank, and version not being set on the json in an existing certificate )

I'll question on some of the other AC and get back to you

CodeWolf86:
Added in the Certification Api Client replacement, Remove the changing of Standard for the privately funded certificates and added in validation for when option isn't selected when changing version.

CodeWolf86 (dismissed):

CodeWolf86 (dismissed):

CodeWolf86 (approved):

PR #259, opened 2021-07-01
SV-928 standard version in epao overview standards table

ben1stone:


CodeWolf86 (approved):

alandburns (commented):
Can you make sure that this does not do fail if there if "Versions" is Null for older Applications

PR #263, opened 2021-07-06
SV-913 Withdrawal Application Overview Page

techtoniq:
In the tech huddle we discussed about showing some text such as "all versions" if the application is to withdraw from the standard completely. Looking at the way the UI interacts with the API this doesn't appear to be a quick win to achieve, so for now it's displaying the version numbers as per the requirement. If the story gets completed before the deadline and there is time remaining, I can come back to enhance it.

### DIFF
--- a/src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
+++ b/src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
@@ -69,7 +69,7 @@
                         var standardAndVersion = Model.StandardName;
                         if (!string.IsNullOrWhiteSpace(Model.StandardVersion))
                         {
-                            standardAndVersion = $"{standardAndVersion}<br>Version {String.Join(",", Model.StandardVersion)}";
+                            standardAndVersion = $"{standardAndVersion}<br>Version {Model.StandardVersion}";
                         }
 
                         <div class="govuk-summary-list__row">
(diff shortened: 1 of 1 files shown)


src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
```diff
@@ -64,11 +64,17 @@
                             <dd class="govuk-summary-list__value">@Model.StandardVersion</dd>
                         </div>
                     }
-                    @if (!string.IsNullOrWhiteSpace(Model.StandardWithReference) && Model.SequenceNo == ApplyConst.STANDARD_WITHDRAWAL_SEQUENCE_NO)
+                    @if (!string.IsNullOrWhiteSpace(Model.StandardName) && Model.SequenceNo == ApplyConst.STANDARD_WITHDRAWAL_SEQUENCE_NO)
```
ben1stone, line 67: The h1 in the prototype is "Withdrawal application overview" can this be set dynamically with the Sequence number?

ben1stone (changes requested):

src/SFA.DAS.AdminService.Web/Views/Application/Sequence.cshtml
CodeWolf86: I think this is driven by what is retrieved from the application in the QnA Api. We've had a few instances of these whereby it's generic because this title also applies for other applications for version, so i think for now we leave it be. 😞

CodeWolf86 (approved):

PR #264, opened 2021-07-06
SV-914 evaluated withdrawal application overview

techtoniq:
Note this needs SV-913 in order to meet the ACs for the entire page


src/SFA.DAS.AdminService.Web/Views/Application/WithdrawalApplication/_WithdrawalApplicationsPartial.cshtml
```diff
@@ -90,7 +90,12 @@
                                                     }
                                                     else if (application.SequenceNo == ApplyConst.STANDARD_WITHDRAWAL_SEQUENCE_NO)
                                                     {
-                                                        @($"{application.StandardName} ({application.StandardReference})")
+                                                        var standardWithVersion = application.StandardName;
+                                                        if (application.Versions.Any())
```
CodeWolf86, line 94: Can this be null? `application.Versions` ?

src/SFA.DAS.AdminService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -258,6 +259,142 @@ public async Task<IActionResult> Assessment(Guid applicationId, int sequenceNo,
...
+                return View(nameof(WithdrawalDateChange), sequenceVm);
+            }
+
+            if (sequenceVm.SequenceNo == ApplyConst.ORGANISATION_WITHDRAWAL_SEQUENCE_NO)
+            {
+                await UpdateOrganisationStandardWithdrawalDate(organisation.EndPointAssessorOrganisationId, null, null, effectiveToDate);
+            }
+            else
+            {
+                await UpdateOrganisationStandardWithdrawalDate(organisation.EndPointAssessorOrganisationId, sequenceVm.StandardReference, sequenceVm.Versions[sequenceVm.CurrentVersionIndex.Value], effectiveToDate);
```
CodeWolf86, line 386: Got a null reference exception here while trying to approve an application for a standard which only had one version, the versions property on the model was null. Worth a double check

This was withdrawing from the brewer standard, and modifying the withdrawal date page

src/SFA.DAS.AdminService.Web/Controllers/Apply/ApplicationController.cs
```diff
@@ -258,6 +259,142 @@ public async Task<IActionResult> Assessment(Guid applicationId, int sequenceNo,
...
+
+            if (sequenceVm.RequestedWithdrawalDate.HasValue)
+            {
+                if (sequenceVm.SequenceNo == ApplyConst.ORGANISATION_WITHDRAWAL_SEQUENCE_NO)
+                {
+                    await UpdateOrganisationStandardWithdrawalDate(organisation.EndPointAssessorOrganisationId, null, null, sequenceVm.RequestedWithdrawalDate.Value);
+                }
+                else
+                {
+                    await UpdateOrganisationStandardWithdrawalDate(organisation.EndPointAssessorOrganisationId, sequenceVm.StandardReference, sequenceVm.Versions[sequenceVm.CurrentVersionIndex.Value], sequenceVm.RequestedWithdrawalDate.Value);
```
CodeWolf86, line 331: Given the bug below found on editing the withdrawal date page, surprised this also didn't error, something to do with the model?

CodeWolf86 (changes requested):
image
Also found the above link which didn't go anywhere so might want to check that.

Given the size of the PR there are only a couple of unit test files that have been added / modified. Some places we could add are:
- The new validator extension you refactored for dates
- The new RegiesterEditOrganisationStandardVersionViewValidator
- The tweaked RegistedEditOrganisationStandardViewModelValidator
- The Application Controller doesn't have any tests currently , so should have some to cover the new actions that have been put in

CodeWolf86 (approved):

PR #265, opened 2021-07-07
SV-930 edit standard version page

ben1stone:


src/SFA.DAS.AdminService.Web/Extensions/ValidatorExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+            if (!int.TryParse(dayString, out var day) || !int.TryParse(monthString, out var month) ||
+                !int.TryParse(yearString, out var year)) return null;
+
+            if (!IsValidDate(year, month, day))
+                return null;
+
+            return new DateTime(year, month, day);
+        }
+
+        public static bool IsValidDate(int year, int month, int day)
```
CodeWolf86, line 19: Can we rework this a bit? Feel like you could simplify to
```
var dateString = $"day/month/year";
if(DateTime.TryParse(dateString,"dd/MM/yy", out var date))
return date;
else
return null
```

src/SFA.DAS.AdminService.Web/Views/Register/EditOrganisation.cshtml
```diff
@@ -267,16 +267,19 @@
...
                             @foreach (var result in Model.RegisterViewOrganisationStandardsViewModel.OrganisationStandards.PaginationViewModel.PaginatedList.Items.OrderBy(x =>
-                           x.StandardCollation.Title))
+                                x.StandardCollation.Title))
                             {
                                 <tr class="govuk-table__row">
-                                    <td class="govuk-table__cell" data-label="Name">@result.StandardCollation.Title</td>
-                                    <td class="govuk-table__cell" data-label="ID">@result.StandardCode</td>
+                                    <td class="govuk-table__cell" data-label="Name">
+                                        @result.StandardCollation.Title <br />
+                                        Version @string.Join(", ", result.StandardVersions.Select(v => v.Version))
```
CodeWolf86, line 280: Don't know id you need a null / empty check on the standardversions.select within the join?

CodeWolf86 (approved):

PR #277, opened 2021-07-19
APR-2601 - UKPRN allow list page

martin-lush:


MarkFCain (dismissed):

src/SFA.DAS.AdminService.Web/Validators/Roatp/AllowedProviders/AddUkprnToAllowProvidersListValidator.cs
```diff
@@ -0,0 +1,46 @@
...
+                    {
+                        if (!int.TryParse(ukprnInput, out var ukprn))
+                        {
+                            context.AddFailure("Enter a valid UKPRN");
+                        }
+                        else if (ukprn < 10000000 || ukprn > 19999999)
+                        {
+                            context.AddFailure("Enter a valid UKPRN using 8 numbers");
+                        }
+                        else if (await _applyApiClient.GetAllowedProviderDetails(ukprn) != null)
```
chrisfoster76, line 28: This code will run synchronously

chrisfoster76 (commented):
Just a comment about async await being ignored in this context

src/SFA.DAS.AdminService.Web/Controllers/Roatp/RoatpAllowedProvidersController.cs
```diff
@@ -0,0 +1,51 @@
...
+            return View("~/Views/Roatp/AllowedProviders/List.cshtml", model);
+        }
+
+        [HttpPost("/Roatp/AllowedProviders")]
+        public async Task<IActionResult> AddUkprn(AddUkprnToAllowedProvidersListViewModel model)
+        {
+            if (!ModelState.IsValid)
+            {
+                model.AllowedProviders = await _applyApiClient.GetAllowedProvidersList(model.SortColumn, model.SortOrder);
+                return View("~/Views/Roatp/AllowedProviders/List.cshtml", model);
```
chrisfoster76, line 43: Rendering a view off a Post will break the back button if subsequently used to return to the page

martin-lush: Have now implemented PRG

src/SFA.DAS.AdminService.Web/Validators/Roatp/AllowedProviders/AddUkprnToAllowProvidersListValidator.cs
martin-lush: In the MVC pipeline it does but the only thing to do would be to use a GetAwaiter().GetResult() which I believe will be done under the hood.

MarkFCain (approved):

MarkFCain (dismissed):

PR #289, opened 2021-08-02
SV-1114 back link bug

techtoniq:


CodeWolf86 (approved):
Approved as a temporary fix in lieu of the SV-1127 tech debt ticket to address overall back link changes.

PR #296, opened 2021-08-16
SV-1142 - Admin service actioning EPAO register withdrawal and automated withdrawal

StuartReeve102:


alandburns (approved):
Tested locally, and looks OK to me.

src/SFA.DAS.AdminService.Web/Views/Application/Returned.cshtml
```diff
@@ -85,7 +85,7 @@
                         <li>We've sent an email to the applicant confirming this approval.</li>
                         @if (Model.SequenceNo == ApplyConst.ORGANISATION_WITHDRAWAL_SEQUENCE_NO)
                         {
-                            <li>You'll need to manually remove this organisation from the register of end-point assessment organisations.</li>
+                            <li>You'll need to manually remove this organisation from the register of end-point assessment organisations on the agreed date.</li>
```
CodeWolf86, line 88: I think this content for withdraw from the register and withdraw from a standard is wrong. Both refer to manually removing the organisation which now shouldn't be needed i think?

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #297, opened 2021-08-25
CON-3466 Added data validation for missing certificate addressees

LACrossley:


ChrisJWoodcock (approved):

ChrisJWoodcock (approved):
Even though bitwise operators are equivalent to logical operators on boolean types it's not good practice to use bitwise operators on booleans - I should reserve bitwise operators for the special cases of manipulating flags

PR #315, opened 2021-11-18
SV-1281 manually add standard versions

ben1stone:


CodeWolf86:
Defunct PR, feature removed

PR #316, opened 2021-11-23
SV-1281 manual add without sv 1290

ben1stone:


ben1stone:
Closed. A new design based on merging instead of manually adding will be implemented instead

PR #334, opened 2022-02-10
APR-2953 Updated data export for removed organisations

raviudari01:


MarkFCain (commented):
Can you please change the "Removed" to a const variable named something like OrganisationStatusRemoved (check if there's a static class that contains this, and use this instead if it exists).  Thanks

src/SFA.DAS.AdminService.Web/AutoMapperProfiles/RoatpOversightOutcomeExportViewModelProfile.cs
```diff
@@ -14,7 +14,7 @@ public RoatpOversightOutcomeExportViewModelProfile()
                 .ForMember(dest => dest.ApplicationReferenceNumber, opt => opt.MapFrom(source => source.ApplicationReferenceNumber))
                 .ForMember(dest => dest.ApplicationSubmittedDate, opt => opt.MapFrom(source => source.ApplicationSubmittedDate))
                 .ForMember(dest => dest.OrganisationName, opt => opt.MapFrom(source => source.OrganisationName))
-                .ForMember(dest => dest.IsOnRegister, opt => opt.MapFrom(source => !string.IsNullOrWhiteSpace(source.ProviderRouteNameOnRegister)))
+                .ForMember(dest => dest.IsOnRegister, opt => opt.MapFrom(source => !string.IsNullOrWhiteSpace(source.ProviderRouteNameOnRegister) && source.OrganisationStatus != "Removed"))
```
MarkFCain: Can you please change the "Removed" to a const variable named something like OrganisationStatusRemoved (check if there's a static class that contains this, and use this instead if it exists). Thanks

MarkFCain (approved):

MarkFCain (approved):

DevenKShah (approved):

PR #355, opened 2022-11-24
CSP-379-TribalRole-Accessing-Manage-Provider-Description

raviudari01:


MarkFCain (approved):

src/SFA.DAS.AdminService.Web/Views/SearchTrainingProvider/Index.cshtml
```diff
@@ -0,0 +1,9 @@
+﻿@using SFA.DAS.AdminService.Web.Controllers
```
DevenKShah, line 1: This view is going to be part of QA site. Remove this.

src/SFA.DAS.AdminService.Web/Controllers/SearchTrainingProviderController.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace SFA.DAS.AdminService.Web.Controllers
+{
+    [Authorize(Roles = Domain.Roles.TribalTeam)]
```
DevenKShah, line 6: This controller is not supposed to be in here. We need to put a link to web qa instead,

DevenKShah (changes requested):

MarkFCain (approved):

PR #358, opened 2023-01-27
QF-973 Generate MI token using Azure Identity when connecting to RoATP V1 API

SawNawDfE:
Do not merge until QF-971 and QF-972 have been merged


src/SFA.DAS.AdminService.Web/Infrastructure/CompaniesHouseApiClient.cs
```diff
@@ -22,7 +22,7 @@ public class CompaniesHouseApiClient:ICompaniesHouseApiClient
             _logger = logger;
 
             _httpClient = new HttpClient { BaseAddress = new Uri(baseUri) };
-            _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", _tokenService.GetToken());
+            _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", _tokenService.GetToken().GetAwaiter().GetResult());
```
SawNawDfE, line 25: This is the quickest way to make it work with the changes to `IRoatpTokenService`.

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpApplyTokenService.cs
```diff
@@ -1,28 +1,29 @@
...
 
 namespace SFA.DAS.AdminService.Web.Infrastructure
 {
     public class RoatpApplyTokenService : IRoatpApplyTokenService
     {
         private readonly IWebConfiguration _configuration;
-        private readonly IHostingEnvironment _hostingEnvironment;
-        public RoatpApplyTokenService(IWebConfiguration configuration, IHostingEnvironment hostingEnvironment)
+        private readonly IHostEnvironment _hostEnvironment;
+        public RoatpApplyTokenService(IWebConfiguration configuration, IHostEnvironment hostingEnvironment)
```
SawNawDfE, line 12: I made this change because `IHostingEnvironment` is now deprecated, and the suggestion is to use `IHostEnvironment`.

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpApplyTokenService.cs
```diff
@@ -1,28 +1,29 @@
...
-            if (_hostingEnvironment.IsDevelopment())
+            if (_hostEnvironment.IsDevelopment())
                 return string.Empty;
 
             var azureServiceTokenProvider = new AzureServiceTokenProvider();
-            var generateTokenTask = azureServiceTokenProvider.GetAccessTokenAsync(_configuration.ApplyApiAuthentication.Identifier);
+            var generateTokenTask = await azureServiceTokenProvider.GetAccessTokenAsync(_configuration.ApplyApiAuthentication.Identifier);
 
-            return generateTokenTask.GetAwaiter().GetResult();
+            return generateTokenTask;
```
SawNawDfE, line 26: This method should run a little faster now that the calling thread is no longer blocked.

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpApplyTokenService.cs
```diff
@@ -1,28 +1,29 @@
 ﻿using Microsoft.Azure.Services.AppAuthentication;
 using Microsoft.Extensions.Hosting;
 using SFA.DAS.AdminService.Settings;
+using System.Threading.Tasks;
 
 namespace SFA.DAS.AdminService.Web.Infrastructure
 {
     public class RoatpApplyTokenService : IRoatpApplyTokenService
```
SawNawDfE, line 8: @ChrisJWoodcock Should this also be upgraded to use Azure Identity, or can we leave it as it is?

ChrisJWoodcock: We can leave it as-is - otherwise we would have to test it and it exists for the other use of the Admin Service to the new Roatp V2 Service

src/SFA.DAS.AdminService.Web/Infrastructure/CompaniesHouseApiClient.cs
ChrisJWoodcock: This is hair-raising - so the presumably the CompaniesHouseApiClient is not used because how could it be used when it's using a token generated to talk to the Roatp V2 Apply service??

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpApplyTokenService.cs
ChrisJWoodcock: So technically by updating to use the async version of the GetToken in the base interface - you have been forced to update the other client(s) adding the await keyword before the calls - question is: Is a regression test good enough to cover that - and that is not code our team maintains...

src/SFA.DAS.AdminService.Web/Infrastructure/IRoatpTokenService.cs
```diff
@@ -1,7 +1,8 @@
-﻿namespace SFA.DAS.AdminService.Web.Infrastructure
+﻿using System.Threading.Tasks;
+namespace SFA.DAS.AdminService.Web.Infrastructure
 {
     public interface IRoatpTokenService
     {
-        string GetToken();
+        Task<string> GetToken();
```
ChrisJWoodcock, line 6: Instead of doing this could we duplicate the <string> GetToken to the IRoatpApplyTokenService and drop the interface inheritance from that (need to check the StructureMap setup) in that way you can update just the IRoatpTokenService which is the code that our team maintains

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpApplyTokenService.cs
SawNawDfE: This should now be reverted as per discussion, please check

src/SFA.DAS.AdminService.Web/Infrastructure/CompaniesHouseApiClient.cs
SawNawDfE: I've raised a backlog item to investigate this: https://skillsfundingagency.atlassian.net/browse/QF-1046

src/SFA.DAS.AdminService.Web/Infrastructure/IRoatpTokenService.cs
SawNawDfE: This has been reverted as discussed

ChrisJWoodcock (approved):

src/SFA.DAS.AdminService.Web/Infrastructure/RoatpTokenService.cs
```diff
@@ -1,35 +1,30 @@
...
-            var tenantId = _configuration.RoatpApiAuthentication.TenantId;
-            var clientId = _configuration.RoatpApiAuthentication.ClientId;
-            var appKey = _configuration.RoatpApiAuthentication.ClientSecret;
-            var resourceId = _configuration.RoatpApiAuthentication.ResourceId;
-
-            var authority = $"https://login.microsoftonline.com/{tenantId}";
-            var clientCredential = new ClientCredential(clientId, appKey);
-            var context = new AuthenticationContext(authority, true);
-            var result = context.AcquireTokenAsync(resourceId, clientCredential).Result;
+            var result = new DefaultAzureCredential().GetTokenAsync(
```
Jordan-fujitsu, line 24: @ChrisJWoodcock if not using async in `GetTokenAsync()` is it worth using `GetToken()` instead?

PR #360, opened 2023-02-16
QF-1000 Migrate from Newtonsoft.Json to System.Text.Json

SawNawDfE:


ChrisJWoodcock (changes requested):
Still seeing a few references to NewtonSoft.Json.Linq

SawNawDfE:
> Still seeing a few references to NewtonSoft.Json.Linq

I've removed two of those 3 references. One was not being used, and the other needed a light refactor before being removed.

The last one in Application.cs will require some thought. I can guess what it's trying to do, but it's working with dynamic data and there aren't any unit tests I can rely on to guarantee that my refactors would be 1:1.

@ChrisJWoodcock

SawNawDfE:
I've renamed the Web.Tests project file and folder to Web.UnitTests, so that they're detected and ran by the pipelines. The PR might interpret this as a large number of additions and deletions

src/SFA.DAS.AdminService.Settings/WebConfiguration.cs
```diff
@@ -1,30 +1,30 @@
-﻿using Newtonsoft.Json;
+﻿using System.Text.Json.Serialization;
 using SFA.DAS.AdminService.Common.Settings;
 
 namespace SFA.DAS.AdminService.Settings
 {
     public class WebConfiguration : IWebConfiguration
     {
-        [JsonRequired] public AzureApiAuthentication AzureApiAuthentication { get; set; }
+        public AzureApiAuthentication AzureApiAuthentication { get; set; }
```
ChrisJWoodcock, line 8: Why is this one no longer required/included?

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Settings/WebConfiguration.cs
SawNawDfE: It was removed by mistake -- I've re-added it now. I've also replaced several instances of the JsonInclude attribute with JsonRequired, since I think that's more equivalent to Newtonsoft's JsonRequired attribute

ChrisJWoodcock (approved):

PR #364, opened 2023-03-07
QF-1143 - Fix User Details Mapping

PeterDavidWilliams:
The upgrade to .net 6.0 broke the automapping, and the replacement extension method doesn't work. As a result, I've replaced the non-functioning extension method on the automapper with explicit mapping and removed the extension file


SawNawDfE (changes requested):
I don't see any existing tests for the [code where these mappings are used](https://github.com/SkillsFundingAgency/das-admin-service/blob/76092cedf7360e04d10be4c309ac475ee9c285b0/src/SFA.DAS.AdminService.Web/Controllers/RegisterUserController.cs#L30-L58), and as a result, we didn't know that the framework upgrades had broken the Automapper logic until it was deployed to pre-prod.

We can prevent this from happening again by adding this test coverage. It looks like this automapper logic is being used in `RegisterUserController.cs`, lines 30 thru 58. The private method `GetUserViewModel` uses the automapper on line 52, then returns a `RegisterViewAndEditUserViewModel`. We should add tests to this controller to ensure that the mappings are working correctly.

SawNawDfE:
Since we're reworking the Automapping logic, we might as well do it on the latest version of that package, which I believe is 12.x.x

SawNawDfE:
> I don't see any existing tests for the [code where these mappings are used](https://github.com/SkillsFundingAgency/das-admin-service/blob/76092cedf7360e04d10be4c309ac475ee9c285b0/src/SFA.DAS.AdminService.Web/Controllers/RegisterUserController.cs#L30-L58), and as a result, we didn't know that the framework upgrades had broken the Automapper logic until it was deployed to pre-prod.
> 
> We can prevent this from happening again by adding this test coverage. It looks like this automapper logic is being used in `RegisterUserController.cs`, lines 30 thru 58. The private method `GetUserViewModel` uses the automapper on line 52, then returns a `RegisterViewAndEditUserViewModel`. We should add tests to this controller to ensure that the mappings are working correctly.
> 
> Another way to test this is refactor that `GetUserViewModel` method into a class, and test that instead of testing the controller. This way the controller can simply return views and perform redirects while all the automapping and view model generation is delegated to the new class. This will improve the separation of concerns and make it slightly easier to test.
> 
> I updated my comment @PeterDavidWilliams

To be clear, the second approach of refactoring the `GetUserViewModel` method into a class is preferable

SawNawDfE:
Sorry closed by accident, finger trouble

src/SFA.DAS.AdminService.Web/Controllers/RegisterUserController.cs
```diff
@@ -18,19 +17,19 @@ public class RegisterUserController : Controller
...
+        private readonly IUserViewModelHelper _userViewModelHelper;
 
-        public RegisterUserController(IContactsApiClient contactsApiClient, IHttpContextAccessor httpContextAccessor, IOrganisationsApiClient organisationsApiClient, IMapper mapper)
+        public RegisterUserController(IContactsApiClient contactsApiClient, IHttpContextAccessor httpContextAccessor, IOrganisationsApiClient organisationsApiClient, IUserViewModelHelper userViewModelHelper)
         {
             _contactsApiClient = contactsApiClient;
             _organisationsApiClient = organisationsApiClient;
-            _mapper = mapper;
+            _userViewModelHelper = userViewModelHelper;
         }
```
SawNawDfE, line 27: It's best to leave the private fields and constructor arguments as they were. Changing the "internals" of this controller could have unintended side effects somewhere else. In this case, they don't need to change.

This controller should declare an instance of the `UserViewModelHelper` and pass its private data as constructor parameters in the 3 places where it's used, before calling `GetUserViewModel()`. That's what it's doing in the unit test.

src/SFA.DAS.AdminService.Web.Tests/Controllers/Register/Given_I_request_edit_standard_version_page.cs
```diff
@@ -53,4 +53,4 @@ public async Task And_submit_valid_updated_dates_Then_return_redirect_to_view_or
             result.ActionName.Should().Be("ViewStandard");
         }
     }
-}
+}
```
SawNawDfE, line 56: Looks like this file had a newline character before this change

SawNawDfE (changes requested):

src/SFA.DAS.AdminService.Web/Helpers/UserViewModelHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿using AutoMapper;
+using SFA.DAS.AdminService.Web.ViewModels.Register;
+using SFA.DAS.AssessorService.Application.Api.Client.Clients;
+using System;
+using System.Threading.Tasks;
+using static StackExchange.Redis.Role;
+
+namespace SFA.DAS.AdminService.Web.Helpers
+{
+    public class UserViewModelHelper : IUserViewModelHelper
```
SawNawDfE, line 10: This interface can go since it's not needed any more. Other than that, looks good to me

ChrisJWoodcock (changes requested):
I agree with splitting this out for testability but I don't like the Helper class name in the Helpers folder this looks like the Orchestrator pattern so add a folder called Orchestrators and name your new class UserViewModelOrchestrator and set it up for DI so you don't need to new an instance of it in several places

ChrisJWoodcock (approved):

src/SFA.DAS.AdminService.Web/Controllers/RegisterUserController.cs
SawNawDfE: Had a chat with Chris about this. We have very similar ideas for the refactor and the difference is mostly terminology. Resolving

src/SFA.DAS.AdminService.Web/Helpers/UserViewModelHelper.cs
SawNawDfE: Had a chat with Chris about this. We have very similar ideas for the refactor and the difference is mostly terminology. Resolving

ChrisJWoodcock (approved):

PR #378, opened 2023-08-23
QF-1571_Withdrawal_journey_changes

ChrisJWoodcock:
…sion.

### DIFF
--- a/src/SFA.DAS.AdminService.Web/Services/AnswerInjectionService.cs
+++ b/src/SFA.DAS.AdminService.Web/Services/AnswerInjectionService.cs
@@ -229,31 +229,29 @@ namespace SFA.DAS.AdminService.Web.Services
 
             var warningMessages = new List<string>();
 
-            // Organisation checks ////////////////////////////////
+            // Organisation checks 
             RaiseWarningIfNoEpaoId(command.EndPointAssessorOrganisationId, warningMessages);
             RaiseWarningIfEpaoIdIsInvalid(command.EndPointAssessorOrganisationId, warningMessages);
 
-            // Standard checks ///////////////////////////////////
+            // Standard checks 
             RaiseWarningIfStandardCodeIsInvalid(command.StandardCode, warningMessages);
 
             var standard = await MapCommandToOrganisationStandardRequest(command);
-
-            //if a withdrawal exists - then versions must exist - update rather than insert in the assessor service
-            var withdrawal = await _applyApiClient.GetWithdrawnApplications(command.OrganisationId, command.StandardCode);
-            if (withdrawal.Count != 0)
-                standard.ApplyFollowingWithdrawal = true;
-
 
             // If we passed basic pre-checks; then validate fully
             if (warningMessages.Count == 0)
             {
                 var validationResponse = await _assessorValidationService.ValidateNewOrganisationStandardRequest(standard);
 
-                if (standard.ApplyFollowingWithdrawal && validationResponse.Errors.Count == 1 &&
+                // When applying for a standard that was previously withdrawn
+                var applyFollowingWithdrawal = await _applyApiClient.GetLatestWithdrawalDateForStandard(command.OrganisationId, command.StandardCode) != null ? true : false;
+
+                // The organisation is expected to be present already
+                if (applyFollowingWithdrawal && validationResponse.Errors.Count == 1 &&
                     validationResponse.Errors[0].Field == "OrganisationId" &&
                     validationResponse.Errors[0].StatusCode == ValidationStatusCode.AlreadyExists.ToString())
                 {
-                    _logger.LogInformation($"Inject standard on Validation Service - versions must exist so bypass warnings. OrganisationId: {command.OrganisationId})");
+                    _logger.LogInformation($"Inject standard failed on Validation Service. OrganisationId: {command.OrganisationId} - Bypass: Standard previously withdrawn)");
                 }
                 else
                 {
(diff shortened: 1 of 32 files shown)


src/SFA.DAS.AdminService.Web/Services/AnswerInjectionService.cs
```diff
@@ -229,31 +229,29 @@ public async Task<CreateOrganisationStandardFromApplyResponse> InjectApplyOrgani
...
             if (warningMessages.Count == 0)
             {
                 var validationResponse = await _assessorValidationService.ValidateNewOrganisationStandardRequest(standard);
 
-                if (standard.ApplyFollowingWithdrawal && validationResponse.Errors.Count == 1 &&
+                // When applying for a standard that was previously withdrawn
+                var applyFollowingWithdrawal = await _applyApiClient.GetLatestWithdrawalDateForStandard(command.OrganisationId, command.StandardCode) != null ? true : false;
+
+                // The organisation is expected to be present already
+                if (applyFollowingWithdrawal && validationResponse.Errors.Count == 1 &&
```
SawNawDfE, line 250:
```suggestion
                if (applyFollowingWithdrawal && validationResponse.Errors.Single() &&
```

src/SFA.DAS.AdminService.Web/Services/AnswerService.cs
```diff
@@ -164,6 +166,39 @@ public async Task<CreateOrganisationStandardCommand> GatherAnswersForOrganisatio
...
+            };
+        }
+
+        public async Task<WithdrawStandardRequest> GatherAnswersForWithdrawStandardForApplication(Guid applicationId)
+        {
+            var application = await _applyApiClient.GetApplication(applicationId);
+            var applicationData = await _qnaApiClient.GetApplicationDataDictionary(application?.ApplicationId ?? Guid.Empty);
+            var organisation = await _apiApiClient.GetOrganisation(application?.OrganisationId ?? Guid.Empty);
+
+            if (application is null || applicationData is null || organisation is null) return null;
```
SawNawDfE, line 192: These lines are a duplicate of lines 171-175. If this is intentional, they can be refactored into a class that returns the three as a DTO. That way we can add some quick tests. (Test coverage is very low in this PR and in the Admin Service in general)

SawNawDfE (changes requested):
Left a few comments

src/SFA.DAS.AdminService.Web/Services/AnswerInjectionService.cs
ChrisJWoodcock: That suggestion is not equivalent .Single() will throw an exception if there is not 1 item - this check is checking if there is 1 item

src/SFA.DAS.AdminService.Web/Services/ApplicationService.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly IQnaApiClient _qnaApiClient;
+
+        public ApplicationService(IApiClient apiClient, IApplicationApiClient applyApiClient, IQnaApiClient qnaApiClient)
+        {
+            _apiApiClient = apiClient;
+            _applyApiClient = applyApiClient;
+            _qnaApiClient = qnaApiClient;
+        }
+
+        public async Task<WithdrawalApplicationDetails> GetWithdrawalApplicationDetails(Guid applicationId)
```
SawNawDfE, line 24: The return type should indicate that a null return is possible. The compiler then marks the calling line as "maybe null", and code analysis tools can make use of this
```suggestion
        public async Task<WithdrawalApplicationDetails?> GetWithdrawalApplicationDetails(Guid applicationId)
```

src/SFA.DAS.AdminService.Web/Services/ApplicationService.cs
```diff
@@ -0,0 +1,63 @@
...
+            return new WithdrawalApplicationDetails
+            {
+                ApplicationId = application.Id,
+                EndPointAssessorOrganisationId = organisation.EndPointAssessorOrganisationId,
+                StandardCode = application.StandardCode.Value,
+                ConfirmedWithdrawalDate = (DateTime)applicationData[nameof(ApplicationData.ConfirmedWithdrawalDate)]
+            };
+        }
+
+        public async Task<ApplicationDetails> GetApplicationsDetails(Guid applicationId)
```
SawNawDfE, line 41: Same here
```suggestion
        public async Task<ApplicationDetails?> GetApplicationsDetails(Guid applicationId)
```

SawNawDfE (changes requested):

SawNawDfE (dismissed):

src/SFA.DAS.AdminService.Web/Services/ApplicationService.cs
SawNawDfE: I just noticed that a few changes to the project file would be needed to make use of nullables. If you decide to do this, you would have to add these lines to the PropertyGroup section of the csproj:

```
<LangVersion>10.0</LangVersion> <!-- 8.0 or higher will work, but .NET 6 supports C# 10-->
<Nullable>enable</Nullable>
```

This should've been done as part of the .NET 6 upgrade. It's unfortunate that our .NET 6 apps are still using the version of C# released with .NET Core 2.0!

ChrisJWoodcock: I'll leave these for now then

SawNawDfE (dismissed):

SawNawDfE (approved):

furqanagwan (dismissed):

chrisfoster76 (dismissed):

debrr07 (approved):

PR #382, opened 2023-10-13
FAI-940 Admin Service InvalidRole/Access Denied Page updated for DfESignIn Users

balaji-j:
Commit Details:
1. Amended Access Denied Page for DfESignIn Users.
2. Unit Testing included with the test coverage for newly added code.
3. DfESignIn Auth PKG upgraded to the Latest v17.1.56

Story: https://skillsfundingagency.atlassian.net/browse/FAI-940


dashton82 (dismissed):

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTest.cs
SawNawDfE, line 1: A minor thing, but please could this be renamed to `HomeControllerTests` (plural) to be somewhat in line with the naming convention for the other test files

src/SFA.DAS.AdminService.Web/Models/Error403ViewModel.cs
```diff
@@ -0,0 +1,27 @@
...
+            if (!string.IsNullOrEmpty(environment) && !environment.Equals("prd", StringComparison.CurrentCultureIgnoreCase))
+            {
+                _integrationUrlPart = "test-";
+            }
+        }
+
+        /// <summary>
+        /// Gets DfESignIn Select service link.
+        /// </summary>
+        public string HelpPageLink => $"https://{_integrationUrlPart}services.signin.education.gov.uk/approvals/select-organisation?action=request-service";
```
ChrisJWoodcock, line 20: This looks really odd - why wouldn't this Url be a configuration property

src/SFA.DAS.AdminService.Web/Controllers/HomeController.cs
```diff
@@ -29,7 +34,7 @@ public IActionResult Error()
 
         public IActionResult InvalidRole()
         {
-            return View();
+            return View(new Error403ViewModel(_configuration["ResourceEnvironmentName"]) { UseDfESignIn = _webConfiguration.UseDfESignIn });
```
ChrisJWoodcock, line 37: I have never seen this done before using resource environment name from the azure configuration, you should be using configuration values from the storage configuration

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web.UnitTests/Models/WhenBuildingError403ViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        [TestCase("")]
+        [TestCase(null)]
+        public void Then_The_HelpPage_Link_Is_Correct_When_Environment_Is_Null(string env)
+        {
+            var actual = new Error403ViewModel(env);
+
+            Assert.That(actual.HelpPageLink, Is.Not.Null);
+            Assert.That(actual.HelpPageLink, Is.EqualTo("https://services.signin.education.gov.uk/approvals/select-organisation?action=request-service"));
```
SawNawDfE, line 33: Same thing for these hardcoded URLs. The tests don't need to worry about the actual URL, only that whatever URL retrieved from the config is being used. You can do this by mocking the `HelpPageLink` property in your view model

SawNawDfE (changes requested):

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTest.cs
```diff
@@ -49,26 +52,48 @@ public void Then_Index_Returns_ViewModel(bool useDfESignIn)
...
 
             //sut
-            var result = controller.Index();
+            var result = _controller.Index();
 
             //assert
             result.Should().NotBeNull();
             var actualResult = (RedirectToActionResult)result;
             actualResult.ControllerName.Should().Be("Dashboard");
             actualResult.ActionName.Should().Be("Index");
```
SawNawDfE, line 74: If one of these asserts fail, the test will abort without running the subsequent asserts. We can avoid this by wrapping this block in an `AssertionScope`
```suggestion
           using (new AssertionScope())
           {
                result.Should().NotBeNull();
                var actualResult = (RedirectToActionResult)result;
                actualResult.ControllerName.Should().Be("Dashboard");
                actualResult.ActionName.Should().Be("Index");
           }
```

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTest.cs
```diff
@@ -49,26 +52,48 @@ public void Then_Index_Returns_ViewModel(bool useDfESignIn)
...
+
+            //sut
+            var result = (ViewResult)_controller.InvalidRole();
+
+            //assert
+            Assert.That(result, Is.Not.Null);
+
+            var actualModel = result.Model as Error403ViewModel;
+            Assert.That(actualModel?.HelpPageLink, Is.EqualTo(helpLink));
+            Assert.That(actualModel?.UseDfESignIn, Is.EqualTo(useDfESignIn));
```
SawNawDfE, line 96: If one of these asserts fail, the test will abort without running the subsequent asserts. We can make all the asserts run regardless of failure by wrapping this block in an `AssertMultiple`

```suggestion
           Assert.Multiple(() => 
           { 
               Assert.That(result, Is.Not.Null);

               var actualModel = result.Model as Error403ViewModel;
               Assert.That(actualModel?.HelpPageLink, Is.EqualTo(helpLink));
               Assert.That(actualModel?.UseDfESignIn, Is.EqualTo(useDfESignIn));
           });
```

SawNawDfE (changes requested):

src/SFA.DAS.AdminService.Web/Models/Error403ViewModel.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            if (!string.IsNullOrEmpty(environment) && !environment.Equals("prd", StringComparison.CurrentCultureIgnoreCase))
+            {
+                _integrationUrlPart = "test-";
+            }
+        }
+
+        /// <summary>
+        /// Gets DfESignIn Select service link.
+        /// </summary>
```
SawNawDfE, line 19: This property is self-explanatory. It's best to leave comments like this out unless they substantially improve the clarity of the code

src/SFA.DAS.AdminService.Web/Models/Error403ViewModel.cs
```diff
@@ -0,0 +1,27 @@
...
+        public string HelpPageLink => $"https://{_integrationUrlPart}services.signin.education.gov.uk/approvals/select-organisation?action=request-service";
+
+        /// <summary>
+        /// Gets or Sets UseDfESignIn config property.
+        /// </summary>
```
SawNawDfE, line 24: This property is self-explanatory. It's best to leave comments like this out unless they substantially improve the clarity of the code

SawNawDfE (changes requested):

src/SFA.DAS.AdminService.Web/Controllers/HomeController.cs
balaji-j: As the ResourceEnvironemtName is a deployment setting and it change based on the current deployment environment, Values are configured in the Pipeline Library, Similar to the configuration["EnvironmentName"]

src/SFA.DAS.AdminService.Web/Models/Error403ViewModel.cs
balaji-j: As the DfE SignIn Service url are most unlikely to get changed ever, we have hardcoded this Url in the Model rather than as the configuration,

balaji-j: Changed the help link from configuration.

src/SFA.DAS.AdminService.Web/Controllers/HomeController.cs
balaji-j: Changed the help link from configuration. So this change is not required anymore.

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTest.cs
```diff
@@ -49,26 +52,48 @@ public void Then_Index_Returns_ViewModel(bool useDfESignIn)
...
             result.Should().NotBeNull();
             var actualResult = (RedirectToActionResult)result;
             actualResult.ControllerName.Should().Be("Dashboard");
             actualResult.ActionName.Should().Be("Index");
         }
+
+        [TestCase("test", "https://test-services.signin.education.gov.uk/approvals/select-organisation?action=request-service", true)]
+        [TestCase("pp", "https://test-services.signin.education.gov.uk/approvals/select-organisation?action=request-service", true)]
+        [TestCase("local", "https://test-services.signin.education.gov.uk/approvals/select-organisation?action=request-service", false)]
+        [TestCase("prd", "https://services.signin.education.gov.uk/approvals/select-organisation?action=request-service", false)]
```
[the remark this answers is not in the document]
balaji-j: Changed the help link from configuration. So this change is not required anymore.

src/SFA.DAS.AdminService.Web.UnitTests/Models/WhenBuildingError403ViewModel.cs
balaji-j: Changed the help link from configuration. So this change is not required anymore.

src/SFA.DAS.AdminService.Settings/IWebConfiguration.cs
```diff
@@ -29,5 +29,7 @@ public interface IWebConfiguration
         /// Gets or Sets the UseDfESignIn property.
         /// </summary>
         bool UseDfESignIn { get; }
```
SawNawDfE, line 31: This comment can be removed since the property is self-explanatory. Also it doesn't set the property

src/SFA.DAS.AdminService.Settings/WebConfiguration.cs
```diff
@@ -31,5 +31,7 @@ public class WebConfiguration : IWebConfiguration
         /// Gets or Sets the UseDfESignIn property.
         /// </summary>
         [JsonRequired] public bool UseDfESignIn { get; set; } = false;
```
SawNawDfE, line 33: This comment can be removed since the property is self-explanatory. Also, the `UseDfESignIn` property in the interface has no setter, whereas this one does

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            };
+
+            //sut
+            var result = _controller.Index();
+
+            //assert
+            result.Should().NotBeNull();
+            var resultModel = result.Should().BeOfType<ViewResult>().
+                Which.Model.Should().BeOfType<HomeViewModel>().Which;
+            resultModel.UseDfESignIn.Should().Be(useDfESignIn);
```
SawNawDfE, line 46: These asserts should be wrapped in a `using (new AssertionScope())` so the test does not abort on first failed assertion and the test results are accurately reported

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        public void Then_Index_Returns_ViewModel(bool useDfESignIn)
+        {
+            //arrange
+            _mockWebConfiguration.Setup(c => c.UseDfESignIn).Returns(useDfESignIn);
+            _controller = new HomeController(_mockWebConfiguration.Object)
+            {
+                ControllerContext = new ControllerContext { HttpContext = new DefaultHttpContext() },
+            };
+
+            //sut
```
SawNawDfE, line 39: I think you meant to write "act" here, since the sut instanced on line 34

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                new Claim("name", "somename")
+            }, "mock"));
+            var httpContext = new Mock<HttpContext>();
+            httpContext.Setup(c => c.User).Returns(authorizedUser);
+            _controller = new HomeController(_mockWebConfiguration.Object)
+            {
+                ControllerContext = new ControllerContext { HttpContext = httpContext.Object }
+            };
+
+            //sut
```
SawNawDfE, line 66: "act"

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        [Test, MoqAutoData]
+        public void When_InvalidRole_Then_ViewIsReturned(string helpLink, bool useDfESignIn)
+        {
+            //arrange
+            _mockWebConfiguration.Setup(x => x.UseDfESignIn).Returns(useDfESignIn);
+            _mockWebConfiguration.Setup(x => x.DfESignInServiceHelpUrl).Returns(helpLink);
+            _controller = new HomeController(_mockWebConfiguration.Object);
+
+            //sut
```
SawNawDfE, line 87: "act"

SawNawDfE (changes requested):

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
balaji-j: Updated the comment

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
balaji-j: Updated the comment

src/SFA.DAS.AdminService.Web.UnitTests/Controllers/Home/HomeControllerTests.cs
balaji-j: Updated the comment

SawNawDfE (approved):
@balaji-j Thanks for making the changes. I have no further comments, it's now up to @ChrisJWoodcock to review and approve the PR

ChrisJWoodcock (approved):

PR #390, opened 2024-01-19
QF-690 Use MI for RoATP API Authentication

ChrisJWoodcock:


src/SFA.DAS.AdminService.Settings/RoatpApiClientConfiguration.cs
```diff
@@ -0,0 +1,11 @@
+﻿using SFA.DAS.Http.Configuration;
+
+namespace SFA.DAS.AdminService.Settings
+{
+    public class RoatpApiClientConfiguration : IManagedIdentityClientConfiguration
```
Jordan-fujitsu, line 5: @ChrisJWoodcock just to double check on this one, should all client have there own configuration for now and maybe in future we can update the ManagedIdentityClientConfiguration.cs in the shared assessor package to use IManagedIdentityClientConfiguration.cs in SFA.DAS.Http package and update client configs to use that?

ChrisJWoodcock: Yes for now - the ManagedIdentityClientConfiguration will be no longer shared when the QnA Client has been moved to the QnA repo - there is ticket for doing that in the backlog. 
The clients having their own configuration is a pattern I can see in most of the repos which use the SFA.DAS.Http - so we should introduce it in assessor

Jordan-fujitsu (dismissed):

PR #404, opened 2025-02-27
P"-2331-FWReprint_NoResults

karen-hanna:


src/SFA.DAS.AdminService.Web/Validators/SearchInputViewModelValidator.cs
```diff
@@ -52,6 +54,11 @@ public SearchInputViewModelValidator()
                         {
                             context.AddFailure("Date", "The date must be a real date");
                         }
+                        else if (date < minimiumDate)
```
ChrisJWoodcock, line 57: I think this should be somewhere else

ChrisJWoodcock (approved):
There is some code here which probably belongs in a different PR

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
```diff
@@ -63,20 +62,34 @@ public async Task<IActionResult> Results(SearchInputViewModel vm, int page = 1)
...
+                        _sessionService.ClearFrameworkSearchRequest();
+                        return RedirectToAction("NoResults",
+                            new
+                            {
+                                FirstName = searchQuery.FirstName,
+                                LastName = searchQuery.LastName,
+                                DateOfBirth = searchQuery.DateOfBirth
+                            });
+                    }
+                    else
```
ChrisJWoodcock, line 80: If it's more than 1 go to the multiple - else stay here - and the next story would go to the details view for a single result

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
karen-hanna: if (frameworkResults.Count == 0)
{
    _sessionService.ClearFrameworkSearchRequest();
    return RedirectToAction("NoResults",
        new
        {
            FirstName = searchQuery.FirstName,
            LastName = searchQuery.LastName,
            DateOfBirth = searchQuery.DateOfBirth
        });
}
else if (frameworkResults.Count > 1)
{
    var searchSessionObject = new FrameworkSearch()
    {
        FirstName = searchQuery.FirstName,
        LastName = searchQuery.LastName,
        DateOfBirth = searchQuery.DateOfBirth,
        FrameworkResults = _mapper.Map<List<FrameworkLearnerSummaryViewModel>>(frameworkResults)
    };

    _sessionService.SessionFrameworkSearch = searchSessionObject;
    return RedirectToAction("MultipleResults");      
}

src/SFA.DAS.AdminService.Web/Validators/SearchInputViewModelValidator.cs
karen-hanna: rebased onto 2331

PR #407, opened 2025-03-07
P2 2333 fw reprint reprint reason

karen-hanna:


src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
```diff
@@ -187,15 +186,21 @@ public async Task<IActionResult> SelectFrameworkLearner(FrameworkLearnerSearchRe
...
+                return RedirectToAction("Index");
+            }
 
-                return View(_mapper.Map<FrameworkLearnerViewModel>(certificateDetails));
+            var certificateDetails = 
+                await _learnerDetailsApiClient.GetFrameworkLearner(sessionModel.SelectedResult.Value);
+
+            _sessionService.UpdateFrameworkSearchRequest((sessionObject) =>
+            {
+                sessionObject.CertificateNumber = certificateDetails.CertificateNumber;
```
ChrisJWoodcock, line 199: I see now that this is being stored so it can be shown on the next page - but on the check page all the details will be required - so why not save all the details into the session by storing 'certificateDetails' which would be better names 'frameworkLearner'

src/SFA.DAS.AdminService.Web/Views/Search/Certificate.cshtml
```diff
@@ -165,7 +165,7 @@
             <h3 class="govuk-heading-m">What do you want to do next?</h3>
 
             <p class="govuk-body">
-                <a class="govuk-link" asp-controller="Search" asp-action="Certificate">Request certificate reprint</a>
+                <a class="govuk-link" asp-controller="Search" asp-action="Reprint">Request certificate reprint</a>
```
ChrisJWoodcock, line 168: This need to be fitted better with the existing code so that it all makes sense together - in the standard journey there is an action ReprintReason which is the reason for reprinting the standard certificate - this should be renamed as StandardReprintReason and then the new one is FrameworkReprintReason

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
karen-hanna: i changed my mind about storing all the details in the session so this gets removed in a later story.  i though it was better only to store data changed by the user in the session and get the certificate details from the api on each page.  so i'll leave this as iot is for now and if you object on the later story i will change it

src/SFA.DAS.AdminService.Web/Views/Search/Certificate.cshtml
karen-hanna: i have renamed the existing standard reprint reason models/views/methods to StandardReprintReason and framework equivalents are now called FrameworkReprintReason.

src/SFA.DAS.AdminService.Web/Views/Search/FrameworkReprintReason.cshtml
```diff
@@ -0,0 +1,142 @@
...
+            <h1 class="govuk-heading-l">
+                @ViewBag.Title
+            </h1>
+            <dl class="govuk-summary-list">
+                <div class="govuk-summary-list__row">
+                    <dt class="govuk-summary-list__key">
+                        Certificate number
+                    </dt>
+                    <dd class="govuk-summary-list__value">
+                        @Model.CertificateNumber
```
ChrisJWoodcock, line 28: This should be the Certificate Reference when it is present - it's not present now because this is the first reprint

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

PR #408, opened 2025-03-07
P2-2345 Add reprint framework address page

karen-hanna:


src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
```diff
@@ -242,26 +242,67 @@ public IActionResult FrameworkReprintReason()
...
+            {
+                sessionObject.SelectedReprintReasons = vm.SelectedReprintReasons;
+                sessionObject.TicketNumber = vm.TicketNumber;
+                sessionObject.OtherReason = vm.OtherReason;
+            });
+
             if (ModelState.IsValid)
             {
-                //Page refreshes, temporarily clearing out the entered data until next story is implemented
+                return RedirectToAction("Address");
```
ChrisJWoodcock, line 254: FrameworkAddress

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
```diff
@@ -242,26 +242,67 @@ public IActionResult FrameworkReprintReason()
...
+                return RedirectToAction("Index");
+            }
+
+            var viewModel = _mapper.Map<FrameworkLearnerAddressViewModel>(sessionModel);
+            return View(viewModel);
+        }
+
+        [HttpPost]
+        [ModelStatePersist(ModelStatePersist.Store)]
+        public IActionResult UpdateAddress(FrameworkLearnerAddressViewModel vm)
```
ChrisJWoodcock, line 279: It is much more usual to call both the Get and Post methods for a screen the same name

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
```diff
@@ -242,26 +242,67 @@ public IActionResult FrameworkReprintReason()
...
-                    sessionObject.TicketNumber = string.Empty;
-                    sessionObject.OtherReason = string.Empty;
+                    sessionObject.AddressLine1 = string.Empty;
+                    sessionObject.AddressLine2 = string.Empty;
+                    sessionObject.TownOrCity = string.Empty;
+                    sessionObject.County = string.Empty;
+                    sessionObject.Postcode = string.Empty;
                 });
-                return RedirectToAction("FrameworkReprintReason");
+                return RedirectToAction("Address");
```
ChrisJWoodcock, line 292: Should use e.g. nameof(FrameworkAddress) as just more reliable coding

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
karen-hanna: updated

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
karen-hanna: renamed

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
karen-hanna: updated

ChrisJWoodcock (approved):

PR #409, opened 2025-03-10
P2-2346 FW reprint Check and submit

karen-hanna:


src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
```diff
@@ -210,48 +214,68 @@ public async Task<IActionResult> FrameworkLearnerDetailsBackAction()
...
+                }
                 _sessionService.ClearFrameworkSearchRequest();
             }
-            
+
             return RedirectToAction("Index");
         }
 
+        [HttpGet]
+        public IActionResult BackAction(string backActionTarget)
```
ChrisJWoodcock, line 232: Which BackAction is this.....

ChrisJWoodcock (changes requested):

src/SFA.DAS.AdminService.Web/Controllers/SearchController.cs
karen-hanna: I'm using this for the 2 screens where the user can go from the Check Answers page.  It clears the flag which is hel in the session and redirects so i have renamed it to ResetBackToCheckAnswersAndRedirect because that's what it does

ChrisJWoodcock (approved):

PR #425, opened 2025-12-29
CSP-2202 Delete roatp admin functions that are migrated

DevenKShah:
- Applied free license for EPPlus package which is required with the new version. 
- The download register now uses the new endpoint in Roatp service, this required some refactoring to generate excel from object collection. Also I have now excluded the two columns `Parent company guarantee` and `Financial track record` from the export as we are not capturing these anymore.


src/SFA.DAS.AdminService.Web/Views/Dashboard/Index.cshtml
```diff
@@ -53,7 +53,7 @@
         <div class="govuk-grid-column-one-half" sfa-show-for-roles="@Roles.RoatpFinancialAssessorTeam" id="roatp-financial-applications">
             <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
                 <a class="govuk-link govuk-link--no-visited-state" href="@Model.RoatpFinanceBaseUrl">
-                    Manage financial applications
+                    Complete financial health assessment
```
raviudari01: Is this display text change part of this ticket ?

raviudari01 (dismissed):

afzal-hussain259 (approved):

raviudari01 (approved):

ChrisJWoodcock (approved):

